PR #4, opened 2020-12-18
BFS-2.1 solutions

spartan-coder90:


Himanshu-Agarwal93 (commented):
Nice solutions. TC of rottenOranges will be O(m*n) since it is a matrix. Try to use directions array. You won't have to type so much

Himanshu-Agarwal93:
Nice solutions. TC of rottenOranges will be O(m*n) since it is a matrix. Try to use directions array. You won't have to type so much

PR #534, opened 2025-06-25
Rotten Oranges bfs and Employee importance dfs

kkavyaGoyal2003:


super30admin:
Strengths:
- The use of a direction array for exploring neighbors is a good practice.

Areas for Improvement:
- The flag variable is unnecessary. Instead, you can increment the time after processing each level of the BFS, similar to the reference solution.
- The queue could store pairs (or a custom class) instead of separate integers for row and column, which would make the code cleaner and less error-prone.
- The condition `if(fresh == 0) return 0;` should be checked before entering the BFS loop to handle cases where there are no fresh oranges initially.

PR #541, opened 2025-07-05
LC 690, 994

asmodaks30:


super30admin:
Strengths:
- The code is clean, readable, and well-structured.
- Good use of data structures like unordered_map and queue.
- Proper handling of edge cases (e.g., when fresh == 0).

Areas for Improvement:
- In the oranges rotting problem, the time calculation should be adjusted to match the reference solution. The student returns time-1, which is incorrect. The reference solution returns the time immediately when fresh count reaches zero during the BFS, which is more accurate.
- The student could add comments to explain the logic, especially for the oranges rotting problem where the time calculation might be confusing.

PR #547, opened 2025-07-20
Solved

sethu95:


super30admin:
- For problem1.java (orangesRotting):
  - The time calculation is slightly off. The time should be returned as `time - 1` only if all oranges are rotten. Currently, it increments time at the start of each level, which is correct, but the final return should be `time` if `fresh == 0` is checked inside the loop (as in the reference solution). Alternatively, you could return `time - 1` after the loop if `fresh == 0`.
  - Consider adding an early exit condition inside the loop when `fresh == 0` to optimize further, as done in the reference solution.
- For problem2.java (employee importance):
  - The solution is well-implemented and correct. However, the `employees` parameter in the `dfs` method is unused and can be removed to clean up the code.
  - The `targetEmployee` variable is declared but never used and can be removed.

Overall, the solutions are well-written and mostly correct, with minor improvements needed in problem1.java.

PR #548, opened 2025-07-21
Done

NUPUR0307:


super30admin:
- Strengths:
  - The approach is well-explained in the comments, making it easy to follow.
  - The use of BFS is appropriate for this problem, and the implementation is mostly correct.
  - The code is clean and follows good practices.
- Areas for Improvement:
  - The time calculation should be corrected to return `time` instead of `time-1` when all oranges are rotten. The current implementation might undercount the time by 1.
  - Consider returning the time as soon as `fresh == 0` during the BFS to optimize further, as done in the reference solution.
  - The condition `if(queue.isEmpty()) return -1;` is redundant because if there are no rotten oranges and `fresh > 0`, the BFS will not run, and the final check `if(fresh != 0) return -1;` will handle this case.

PR #555, opened 2025-08-10
submitting BFS-2-1

akshay4121:


super30admin:
- **Strengths**: 
  - The BFS approach is correctly implemented.
  - Proper handling of edge cases (e.g., no fresh oranges initially).

- **Areas for Improvement**:
  - The "rotted" flag is unnecessary. Time should be incremented after processing each level in the BFS, regardless of whether any oranges were rotten in that level.
  - The helper method could be merged into the main method for simplicity, as it doesn't significantly improve modularity in this case.
  - The return statement could be simplified by directly returning `time` if `fresh == 0` and `-1` otherwise, without the ternary operator.

PR #556, opened 2025-08-20
Completed BFS-2-1

eyeamkd:


super30admin:
- Strengths:
  - Correct implementation of BFS to spread the rot.
  - Good use of directions for neighbor checks.
  - Clear and readable code structure.

- Areas for Improvement:
  - Remove the unnecessary condition `if res == -1: return 0`. This condition is incorrect and not needed.
  - Consider tracking the number of fresh oranges to return early if all are rotten, similar to the reference solution.
  - Rename the variable `infected` to something more descriptive like `hasInfected` or `hasSpread`.

PR #557, opened 2025-08-22
Done BFS-2-1

akhilsai25:


super30admin:
- The solution is well-implemented and follows best practices. The use of a queue for BFS and a counter for fresh oranges is appropriate.
- The early termination condition (`if(freshOranges==0) return minutes;`) is a good optimization.
- The code could benefit from a minor improvement in the final return statement. Instead of checking `if(freshOranges!=0) return -1;` followed by `return minutes;`, it could directly return `-1` if `freshOranges` is not zero, making the logic slightly clearer.
- The comments are helpful but could be more detailed to explain the BFS process and the significance of the `minutes` counter.

PR #558, opened 2025-08-22
completed BFS-2-1

DhanyasriG:


super30admin:
Strengths:
- The BFS implementation is clear and correctly handles the traversal.
- The use of a queue for BFS and a map for quick access in Problem2.java is efficient and well-implemented.

Areas for Improvement:
- In Problem1.java, the time calculation could be improved. Instead of returning time-1, consider incrementing the time at the start of each level or adjusting the logic to ensure the correct time is returned without subtraction.
- Ensure edge cases are handled correctly, such as when there are no fresh oranges initially or when all oranges are rotten in the last level.

PR #561, opened 2025-08-23
completed bfs-2-1

nitishred13:


super30admin:
- Strengths:
  - The student demonstrates a good understanding of BFS and its application to both problems.
  - The code is well-organized and includes comments explaining the approach.
  - The use of data structures like HashMap and Queue is appropriate and efficient.
- Areas for Improvement:
  - In the RottingOranges solution, the time increment logic should be adjusted to match the reference solution. The time should be incremented after processing each level of the BFS, not before.
  - The early termination condition (`if(fresh == 0) return time;`) should be checked inside the inner loop to immediately return when all oranges are rotten, as done in the reference solution.
  - The variable names in the RottingOranges solution could be more descriptive (e.g., `temp` could be `current`).

PR #564, opened 2025-09-02
Done BFS-2-1

AneerudhS27:


super30admin:
- The use of comments to explain the logic is helpful and demonstrates good coding practices.
- The early termination condition within the loop is a good optimization.
- The commented-out line can be removed as it is redundant and not needed.
- Overall, the code is clean, efficient, and easy to understand.

PR #574, opened 2025-09-22
BFS-2-1 Complete

dhpatel860:


super30admin:
- The student's solution is well-implemented and follows best practices. The comments are helpful in understanding the approach and complexity.
- One minor improvement could be to move the `time++` increment to after the level processing is complete, similar to the reference solution, to ensure the time is incremented only after all oranges at the current level have been processed. However, the current implementation still works correctly.
- The student could also consider adding a check for `fresh == 0` before entering the BFS loop, similar to the reference solution, to handle the case where there are no fresh oranges initially more elegantly.

PR #578, opened 2025-09-23
BFS 2 -1 Done

desaiparam:


super30admin:
Strengths:
- The comments clearly explain the intended approach.
- The student has correctly identified the time and space complexity.

Areas for Improvement:
- The problem requires a BFS approach to ensure the minimum time to rot all oranges. DFS is not suitable here as it does not guarantee the minimum time.
- The student should reconsider the algorithm choice and implement BFS instead of DFS.
- The student should also ensure that the solution correctly handles edge cases, such as when there are no fresh oranges initially.

PR #583, opened 2025-10-02
BFS-2-1

v-hemish:


super30admin:
- For the oranges rotting problem (problem1.py):
  - Start with `time = 0` instead of `time = 1` to correctly calculate the minutes.
  - Remove the unnecessary condition `if time > 0: return -1` at the end. The correct condition should be `if fc == 0: return time else return -1`.
  - The logic for handling the queue and processing each level is correct and matches the reference solution.

- For the employee importance problem (problem2.py):
  - The solution is well-structured and correct. The use of DFS is appropriate here.
  - Consider using BFS as an alternative approach, which might be more intuitive for some readers.

Overall, the student has done a good job, but there are minor improvements needed in the oranges rotting problem to make it fully correct and efficient.

PR #585, opened 2025-10-08
Completed

shantanujoshi25:


super30admin:
Strengths:
- Both solutions are logically sound and solve the problems correctly.
- The code is readable and well-structured.
- Good use of helper functions and data structures where needed.

Areas for Improvement:
- In Problem_1.py, consider using a direction array (like in the reference solution) to make the code more concise and easier to maintain.
- The time calculation in Problem_1.py could be more accurate if you increment the time after processing all oranges at the current level, rather than before.
- For Problem_2.py, while the DFS approach is correct, mentioning that BFS could also be used would show a broader understanding of the problem.

PR #586, opened 2025-10-11
solved these problems

jeetusingh247:


super30admin:
Strengths:
- The BFS approach is correctly implemented, and the logic for spreading the rot is sound.
- The use of a queue for BFS and tracking fresh oranges is efficient and correct.

Areas for Improvement:
- The time calculation has a minor flaw. The `time` variable should not be decremented by 1 when returning the result. The BFS correctly increments `time` after processing each level, so returning `time` directly is accurate.
- Consider returning early when `fresh` becomes 0 during the BFS traversal to optimize the solution further, as shown in the reference solution.

jeetusingh247:
> Strengths:
> 
> * The code is clean, well-commented, and easy to understand.
> * The BFS approach is correctly implemented, and the logic for spreading the rot is sound.
> * The use of a queue for BFS and tracking fresh oranges is efficient and correct.
> 
> Areas for Improvement:
> 
> * The time calculation has a minor flaw. The `time` variable should not be decremented by 1 when returning the result. The BFS correctly increments `time` after processing each level, so returning `time` directly is accurate.
> * Consider returning early when `fresh` becomes 0 during the BFS traversal to optimize the solution further, as shown in the reference solution.

thankyou for the feedback, I have made the changes !!

PR #589, opened 2025-10-27
Completed BFS-2-1

nikhilsaimarri:
Implement BFS to rot fresh oranges in a grid.


super30admin:
- The use of comments to explain the approach is commendable and helps in understanding the code.
- The variable names are appropriate and make the code readable.
- One minor improvement could be to add a comment explaining the direction vectors (dirs) for clarity, although their usage is straightforward.
- The solution handles edge cases effectively, such as when there are no fresh oranges initially.

PR #591, opened 2025-11-01
BFS-2-1 solutions

kalyan-ng22:


super30admin:
Strengths:
- The use of a directions array for neighbor traversal is a good practice.
- The solution efficiently checks for the condition when all fresh oranges are rotten and returns early.

Areas for Improvement:
- The variable names `topDown` and `leftRight` could be more descriptive, such as `row` and `col` to match conventional grid traversal terminology.
- The comment "each level traversal" could be more specific, such as "explore all four directions from the current cell."

PR #595, opened 2025-11-10
BFS-2-1 Completed

PrasiddhShah:


super30admin:
- Strengths:
  - Good use of BFS for both problems.
  - Clear and concise code with appropriate comments.
  - Correct handling of edge cases, such as when there are no fresh oranges initially.
- Areas for Improvement:
  - In the first problem, consider incrementing the time only after processing all nodes at the current level, but before processing the next level. This would make the time calculation more intuitive and eliminate the need to subtract 1 at the end.
  - The comment in the first problem mentions "mangos" instead of "oranges," which is a minor inconsistency.
  - The time complexity comments in both problems could be more specific. For example, in the first problem, it should be O(m*n) instead of O(n).

PR #600, opened 2025-12-21
Solved BFS-2-1

Dheepthi-Reddy:
- Solved Rotting Oranges
- Solved Symmetric Tree


super30admin:
- Strengths:
  - Both solutions are correct and efficient, matching the expected time and space complexities.
  - Good use of BFS for the oranges problem and DFS for the tree symmetry problem.

- Areas for Improvement:
  - In "RottingOranges.py", consider incrementing the time after processing each level to avoid the need for the final adjustment (time - 1). This would make the logic more straightforward.
  - In "SymmetricTree.py", the use of a class variable (self.flag) is not necessary. You could return boolean values directly from the DFS function to simplify the code.

PR #604, opened 2026-01-23
BFS 2.1

praxpk:


super30admin:
Strengths:
- The student correctly identifies the BFS approach, which is optimal for this problem.
- The code is structured and readable, with clear variable names and comments.

Areas for Improvement:
- The visited set is redundant. Instead, you can directly update the grid to mark oranges as rotten when they are processed, which would simplify the code and save space.
- The initial max_time should be set to 0, not -1, to correctly handle cases where no time is needed (no fresh oranges initially).
- The condition to check if fresh_oranges > 0 or max_time == -1 is not entirely correct. It should only check if fresh_oranges > 0 to determine if it's impossible to rot all oranges.

Suggested Improvements:
- Remove the visited set and update the grid directly when an orange is rotten.
- Initialize max_time to 0 and adjust the logic to return 0 if there are no fresh oranges initially.
- Simplify the final condition to only check if fresh_oranges > 0.

PR #606, opened 2026-01-25
BFS-2-1 Solutions

dbkuppagiri:


super30admin:
Strengths:
- The BFS solution is well-implemented and follows the standard approach. You used a queue and processed levels correctly, which is efficient.
- You handled the directions array and boundary checks properly.
- You kept track of fresh oranges and returned the time when no fresh oranges remain.

Areas for Improvement:
- The DFS solution for the oranges rotting problem is not appropriate. BFS is the natural choice for problems that require minimum steps or propagation in layers. Avoid using DFS for such problems.
- In the DFS solution, you attempted to store the time directly in the grid, but this approach is error-prone because it overwrites the original values and makes it hard to track. Moreover, without a way to ensure the minimum time from any source, it can yield incorrect results.
- The BFS solution could be slightly optimized by breaking early when freshOrangeCount becomes zero, which you did correctly.
- The code for the DFS solution has a potential issue: the initial call uses time=2, but if there are multiple rotten oranges, they might be processed in an arbitrary order, and the DFS might not update adjacent cells with the minimum time.

Recommendation:
Stick to the BFS solution for this problem. The DFS solution should be discarded as it is not suitable. For future problems, remember that BFS is used for shortest path problems in unweighted graphs (or grids), while DFS is better for traversal without the requirement of minimum steps.

PR #607, opened 2026-01-25
completed BFS-2-1

kkhevna9252:


super30admin:
Your solution for the "Oranges getting rotten" problem is well-structured and follows the BFS approach correctly. However, there is a key issue with the time calculation:

1. You are incrementing `time` at the start of processing each level of the BFS. This means that after the initial rotten oranges (minute 0), you set `time` to 1 and then process the first minute. But when you rot the last fresh orange in a level, you return the current `time`, which is one minute more than the actual time taken. For example, in the first example, it should take 4 minutes, but your solution would return 5. To fix this, you should not increment `time` at the beginning of the while loop for the initial rotten oranges. Instead, you can initialize `time` to 0 and then only increment after processing each level, but return `time` only when all fresh oranges are rotten. Alternatively, you can adjust the return value by returning `time-1` when fresh becomes zero, but that might be less intuitive.

   The reference solution handles this by starting with `time=0` and then incrementing at the beginning of each level after the first. However, it returns the current `time` when fresh becomes zero during the processing of that level. In your code, you are doing the same, but note that the initial rotten oranges are at minute 0. So when you start, you have the initial queue. Then you enter the while loop and immediately increment `time` to 1. Then you process the initial rotten oranges. This means that the first minute is actually being counted before any propagation? Actually, the initial rotten oranges are already rotten at time 0. The first minute of propagation should happen after minute 0, and after that minute, we are at minute 1. But the problem requires the minimum minutes until no fresh orange remains. So if we start at minute 0 and after one minute we have rotted some oranges, then the time elapsed is 1.

   However, in your code, when you process the initial queue (which is at minute 0), you increment `time` to 1 and then process the neighbors. This is incorrect because the initial state is at minute 0. You should not increment the time until after you have processed the first level of propagation. One common approach is to set `time = -1` initially and then set it to 0 if there are no fresh oranges. But the reference solution does it by having `time=0` initially and then only incrementing after processing each level, but it returns the `time` when fresh becomes zero during the processing of a level.

   Actually, let's compare with the reference solution:
   - The reference solution has `time = 0` initially.
   - It checks if there are no fresh oranges, returns 0.
   - Then it enters the while loop. Inside, it does `time++` and then processes the current queue (which contains the rotten oranges at the current time). So for the initial rotten oranges, the first time it enters the loop, it sets `time` to 1 and processes the initial rotten oranges. But wait, the initial rotten oranges are already in the queue at time 0. So the reference solution is actually processing the next minute (minute 1) when it first increments time.

   However, the reference solution is correct because it only increments the time when it starts processing the next level. The key difference is that the reference solution does not consider the initial rotten oranges as a minute that causes rotting. The initial state is minute 0. Then, after one minute, the neighbors become rotten. So the reference solution returns the number of minutes that have passed.

   In your code, you are doing the same: you have `time=0` initially, and then when you enter the while loop, you increment `time` to 1 and process the current queue. This is correct for the first minute. However, if there are no fresh oranges to rot (i.e., the initial queue has rotten oranges but no adjacent fresh), then you would still increment time to 1 and then return -1? But actually, if there are no fresh oranges initially, you return 0.

   But the problem is that when you rot the last orange during the processing of a level, you return the current `time`. This is correct because that minute has just completed. For example, if it takes 4 minutes to rot all, then after processing the fourth minute, you return 4.

   However, wait: in the reference solution, the code is:

        while(!q.empty()){
            int size = q.size();
            time++;
            for(int i = 0; i < size; i++){
                ... // process
                if(fresh == 0) return time;
            }
        }

   So it returns the time that was just incremented. This is correct because the minute that caused the last orange to rot is the current minute.

   But in your code, you are doing exactly the same. So why would it be wrong? Actually, it should be correct. Let me test with the example:

   Example 1: 
        grid = [[2,1,1],[1,1,0],[0,1,1]]

   Initially, fresh = 4.
   Queue has one rotten orange at (0,0).

   First iteration of while:
        time becomes 1.
        Process the one rotten orange: it has neighbors (0,1) and (1,0). Both become rotten. fresh becomes 2.
        Add (0,1) and (1,0) to queue.

   Second iteration:
        time becomes 2.
        Process (0,1): it has neighbors (0,2) and (1,1). (0,2) becomes rotten (fresh becomes 1), and (1,1) becomes rotten (but wait, (1,1) was fresh? Actually, (1,1) is fresh initially. So we rot (0,2) and (1,1). But note: when processing (0,1), it will rot (0,2) and (1,1). Then when processing (1,0), it will try to rot (1,1) again but it is already rotten.
        So after processing the two, fresh becomes 1? Actually, we need to see the order.

   Actually, the reference solution would process both (0,1) and (1,0) in the second minute. Then during the second minute, it will rot (0,2) and (1,1) from (0,1), and from (1,0) it will rot (1,1) (already rotten) and (2,0) (which is empty). So after minute 2, fresh is 1 (at (2,1)).

   Then third iteration:
        time becomes 3.
        Process the queue: which has (0,2) and (1,1). 
        From (0,2): it has neighbors (0,3) (out of bound), (1,2) (which is 0), (0,1) (already rotten), (0,3) (out). So no new.
        From (1,1): it has neighbors (1,0) (already rotten), (1,2) (0), (0,1) (rotten), (2,1) (fresh). So we rot (2,1). fresh becomes 0. Then return time=3.

   But the expected output is 4.

   Wait, this indicates that the reference solution I provided might be incorrect? Or my analysis is wrong.

   Actually, the reference solution I provided is in C++ and it is correct. The issue is that the initial queue has only one rotten orange? In the example, there is only one initially? Actually, the grid has:

        [2,1,1]
        [1,1,0]
        [0,1,1]

   So there is one rotten orange at (0,0). But wait, that is not true: there is only one rotten orange at the start.

   However, the reference solution works correctly. The difference is that in the reference solution, the time is incremented at the beginning of the while loop, and then it processes the queue. So for the first minute (time=1), it processes the initial rotten orange. Then for the second minute (time=2), it processes the oranges rotted in minute 1. Then for the third minute (time=3), it processes the oranges rotted in minute 2. And so on.

   But in the example, it takes

PR #608, opened 2026-01-26
Add files via upload

vaishnavi2231:


super30admin:
Strengths:
- You provided two solutions for the problem, showing your ability to think about different approaches.
- The BFS solution is well-implemented and follows the standard approach for multi-source BFS.
- Your code is clean and well-commented.

Areas for Improvement:
- The DFS solution for "Oranges Getting Rotten" is not correct. This problem is typically solved with BFS because we need the minimum time for propagation, which BFS naturally provides (since it processes level by level). DFS is not suitable because it does not guarantee the minimum time when there are multiple paths. I recommend removing the DFS solution for this problem to avoid confusion.
- In the BFS solution, you might consider initializing `time = -1` and then incrementing after processing each level, so that the initial state (minute 0) is represented correctly. However, your current approach (starting with time=0 and then incrementing at the start of the while loop) is also correct because you return time only when fresh becomes zero during processing. But note: if there are no rotten oranges initially, your code would not enter the while loop, and then return -1 if there are fresh oranges? Actually, in your code, if there are no rotten oranges initially, the queue is empty, so the while loop is skipped. Then if fresh>0, it returns -1. This is correct. So your BFS is correct.
- However, in the BFS solution, you should check if there are no fresh oranges at the beginning and return 0 immediately. You do that, which is good.
- For the DFS solution, it is important to understand that not every problem is suitable for DFS. When we need the shortest path or minimum time in an unweighted grid, BFS is the right choice.

PR #622, opened 2026-03-06
Complete BFS-2-1

PrakarshKamal:


super30admin:
Your solution for the "Rotting Oranges" problem has the right idea (using BFS) but contains a few critical bugs and missing checks.

Strengths:
- You correctly identified the BFS approach for simulating the rotting process.
- You used a queue to process the rotten oranges level by level.
- You maintained a count of fresh oranges to check if all become rotten.

Areas for improvement:
1. **Boundary Check**: In the condition for checking the next cell, you have `c < n` but it should be `c < m` because `m` is the number of columns. This is because you defined `n = grid.length` (rows) and `m = grid[0].length` (columns). So the correct condition should be:
   `if (r >= 0 && c >= 0 && r < n && c < m && grid[r][c] == 1)`
   This error might cause an ArrayIndexOutOfBoundsException.

2. **Handling No Fresh Oranges Initially**: You should check if `freshCount` is zero before starting the BFS and return 0 immediately. This handles the case when there are no fresh oranges at minute 0. Currently, your code does not handle the case when there are no rotten oranges and no fresh oranges: it would return -1 (because after the while loop, `time` is 0, and you return `time-1` which is -1) but it should return 0.

3. **Early Termination**: During the BFS, when you rot an orange, you decrement `freshCount`. You should check immediately after decrementing if `freshCount` becomes 0. If it does, you can return the current time + 1? But note: you haven't incremented `time` for the current level yet. Actually, you are processing a level. The time for the current level is the current value of `time` (which is the time for the previous level) plus 1. However, you increment `time` at the end of the level. So if you are processing the first level (which is the initial rotten oranges), `time` is 0. Then you process the neighbors and for each fresh orange you rot, you decrement `freshCount`. If `freshCount` becomes 0 during this level, then the minutes required is 1. But after processing this level, you increment `time` to 1. So you should return `time` (which will be 1) immediately when `freshCount` becomes 0. However, you cannot return inside the inner for-loop because you are iterating over the directions. Instead, you should check after processing all directions for one orange? Actually, you should check after each orange you rot. So you can break early.

   The reference solution does:
        for (each neighbor) {
            if (fresh orange) {
                rot it;
                fresh--;
                if (fresh==0) return time;
            }
        }
   But note: the reference solution increments `time` at the beginning of the while loop (so the first level returns time=1). However, the reference solution initializes `time=0` and then if there are no fresh, returns 0. Then in the while loop, it sets `time++` at the start. So for the first level, time becomes 1.

   In your code, you have:
        int time = 0;
        while (!q.isEmpty()) {
            int size = q.size();
            time++;   // you increment here -> so for the

PR #625, opened 2026-03-08
working solution

avcode3:


super30admin:
Your solution for the "orangesRotting" problem is mostly correct in terms of logic. You have implemented a BFS approach which is suitable for this problem. However, there are a few areas for improvement:

1. **Efficiency of the Queue**: You are using a list as a queue and using `pop(0)` to remove the first element. This operation has O(n) time complexity because it requires shifting all the remaining elements. For larger grids, this can be inefficient. Instead, you should use `collections.deque` which provides O(1) time complexity for appending and popping from both ends. For example:
   ```python
   from collections import deque
   my_queue = deque()
   # Then use my_queue.append() to add and my_queue.popleft() to remove.
   ```

2. **Return Statement**: Your code returns -1 at the end if the while loop exits. This is correct. However, note that in your code, the variable `day_count` is incremented at the start of each minute. This is correct because the first minute (when we process the initial rotten oranges) is counted as minute 1. But note that if there are no rotten oranges initially and there are fresh oranges, you return -1 correctly. Also, if there are no fresh oranges initially, you return 0 correctly.

3. **Edge Cases**: Your code handles edge cases correctly, such as no fresh oranges or no rotten oranges.

4. **Code Clarity**: Your code is readable and well-structured. The variable names are clear. However, you should add comments to explain the steps.

5. **Minor Issue**: You have a variable named `good_count` which actually represents the number of fresh oranges. It might be better to name it `fresh_count` for clarity.

To improve your solution, replace the list with a deque. Also, consider adding comments to explain the steps.

Here is a revised version of your code using deque:
```python
from collections import deque

class Solution:
    def orangesRotting(self, grid: List[List[int]]) -> int:
        m = len(grid)
        n = len(grid[0])
        fresh_count = 0
        queue = deque()
        
        for i in range(m):
            for j in range(n):
                if grid[i][j] == 2:
                    queue.append((i, j))
                elif grid[i][j] == 1:
                    fresh_count += 1
        
        if fresh_count == 0:
            return 0
        
        minutes = 0
        directions = [(0,1), (1,0), (0,-1), (-1,0)]
        
        while queue:
            minutes += 1
            for _ in range(len(queue)):
                i, j = queue.popleft()
                for dx, dy in directions:
                    r, c = i + dx, j + dy
                    if 0 <= r < m and 0 <= c < n and grid[r][c] == 1:
                        grid[r][c] = 2
                        fresh_count -= 1
                        queue.append((r, c))
                        if fresh_count == 0:
                            return minutes
        
        return -1
```

Note: In this revised code, we increment `minutes` at the start of each minute (each level of BFS). This is correct because the first level of BFS (processing the initial rotten oranges) will cause adjacent fresh oranges to become rotten in minute 1. Also, we return `minutes` when fresh_count becomes zero during the processing.

However, note that the initial rotten oranges are already in the queue. The first iteration of the while loop (with minutes=1) processes these initial rotten oranges. This is correct because the rotting process starts immediately.

But wait: if we have initial rotten oranges, they are already rotten at minute 0. Then at minute 1, the oranges adjacent to them become rotten. So the first minute that we count is minute 1. So the code is correct.

PR #628, opened 2026-03-15
Add implementation for leetcode problems 994, 690

rishigoswamy:


super30admin:
### Oranges getting rotten (leetcode_690.py)
It appears you have submitted a solution for the wrong problem. The problem you solved is "Employee Importance" (LeetCode 690), but the assigned problem is "Oranges Rotting" (LeetCode 994). Please ensure you are working on the correct problem.

For the "Oranges Rotting" problem, you need to:
1. Count the number of fresh oranges initially.
2. Use a BFS starting from all rotten oranges (multiple sources) to simulate the rotting process minute by minute.
3. Track the minutes taken and decrement the fresh count as oranges become rotten.
4. After BFS, if there are still fresh oranges left, return -1; otherwise, return the minutes.

You can refer to the reference solution provided for the correct approach. Note that the grid uses 0 (empty), 1 (fresh), and 2 (rotten), and the adjacency is 4-directional.

VERDICT: NEEDS_IMPROVEMENT

---

### Importance of Employee (leetcode_994.py)
It seems there was a misunderstanding: the problem you were asked to solve is "Importance of Employee" (LeetCode 690), not "Rotting Oranges" (LeetCode 994). Your solution is for the wrong problem.

For the "Importance of Employee" problem, you need to calculate the total importance of an employee and all their direct and indirect subordinates. The typical approach is to use a BFS or DFS starting from the given employee ID, after creating a mapping from employee ID to the employee object for quick lookup.

Here are some tips for the correct problem:
1. Create a map (dictionary) to quickly access any employee by their ID.
2. Use a queue (for BFS) or a stack (for DFS) to traverse the hierarchy.
3. Start with the given ID, add its importance, and then process all subordinates recursively.

Your current solution for "Rotting Oranges" is not relevant to this problem. Please revisit the problem statement and implement the solution accordingly.

VERDICT: NEEDS_IMPROVEMENT

PR #632, opened 2026-05-11
completed bfs-2-1

pratikb0501:


super30admin:
### Oranges getting rotten (Problem_1.py)
**Strengths:**
- Correct BFS approach for the problem
- Good use of Python's deque for queue operations
- Proper handling of edge cases (no fresh oranges)
- Clean variable naming and reasonable structure

**Areas for Improvement:**
1. **Time tracking bug**: The early return `return time + 1` inside the inner loop is confusing. The time should be incremented after processing each "minute" of oranges, not before. Consider moving the time increment to the beginning of each BFS level or restructuring to make the logic clearer.

2. **Final return logic**: The `return time - 1` at the end is confusing and potentially incorrect. The reference solution handles this more cleanly by returning `time` when fresh becomes 0 during BFS, and returning -1 if fresh > 0 after BFS completes.

3. **Simplify the logic**: The reference solution's approach is cleaner - increment time at the start of each BFS round, then return time when fresh becomes 0. This makes the time tracking more intuitive.

**Suggested improvement:**
```python
while q:
    time += 1
    qs = len(q)
    for _ in range(qs):
        cr, cc = q.popleft()
        for x, y in neighbours:
            nr, nc = cr + x, cc + y
            if 0 <= nr < rows and 0 <= nc < cols and grid[nr][nc] == 1:
                fresh -= 1
                grid[nr][nc] = 2
                q.append((nr, nc))
    if fresh == 0:
        return time
return -1
```

VERDICT: NEEDS_IMPROVEMENT

---

### Importance of Employee (Problem_2.py)
**Strengths:**
- Correct BFS implementation that properly handles the tree/graph traversal
- Good variable naming (adj, importance, result)
- Properly handles edge cases (employee with no subordinates)
- Logic is sound and follows the same approach as the reference solution

**Areas for Improvement:**
1. Remove the debug `print(adj)` statement before final submission
2. Consider using `collections.defaultdict(list)` for cleaner adjacency list construction
3. The initial check `if id not in adj` is unnecessary - the code would work fine without it since the while loop simply wouldn't execute if there are no subordinates

VERDICT: PASS

